PR #102, opened 2016-01-27
Remove work_poster template

Elarnon:
The work_poster template was included a lot in works list pages
(e.g. Anime list or recommendations), and was a very small wrapper
around work_rating. To improve performance and essentially cut in half
the number of templates to render on these pages, this removes
work_poster entirely in favor of inlining it in the 3 places where it
was used.

Subsequent patches should reduce these 3 places to only 2 or even 1.


Elarnon:
It's actually easier to do all of this in a big patch, closing.

PR #110, opened 2016-01-28
Add category_id in Work

Elarnon:
This adds a new category_id field in mangaki.models.Work, that will be
used to represent whether the Work is an Anime, a Manga or an Album
(or perhaps more ine the future!).

The synchronisation with the existence of an appropriate
Anime/Manga/Album record is done in the Work.save method, on object
creation.

We use an integer field for performance reasons (we will probably want
to filter on this field in the future, thus we want an efficient
index), but the `Work.category' property abstracts this
away.

Moreover, `Work' now uses a custom manager with a convenience
`category' method, such that all Works that are animes can be accessed
through `Work.objects.category('anime')`.


Elarnon:
Mis à jour avec une vraie table `Category` suite à discussion sur le Slack.

PR #111, opened 2016-01-28
Store the rankings in the database

Elarnon:
For performance reasons, we can't compute the works' rankings whenever
a sorting is required (i.e. on an AnimeList or MangaList
page). Previously, the solution for that was using lists of Work IDs
stored as strings in the database, using the Deck table. This patch
completely replaces the Deck table in favor of using (indexed) caching
fields on the Work table directly.

In order to be able to more easily test this patch without damaging
the database, and even though they are not used anymore, this does
*not* remove the Deck table - applying, then inverting the migration
from this patch will be an identity operation on your database.

(Depends on #110)


Elarnon:
Mis à jour pour utiliser la nouvelle table `Category`

PR #112, opened 2016-01-28
Delete the Deck table

Elarnon:
After #111, the Deck table is no longer used and can be safely removed.

(Depends on #111)


Elarnon:
Mis à jour pour utiliser la nouvelle table `Category`.

PR #128, opened 2016-04-09
Add a generic relation between Work and Artist

Elarnon:
Previously, we were using ad-hoc fields on the descendant classes of
Work for their relations with artists (e.g. Anime.author,
Manga.mangaka) that was getting tedious.

As a helper towards the unification of the Anime/Manga/Album/Work
models, this patch replaces those ad-hoc fields with a generic
ManyToMany relation between Works and Authors with information about
the artist's role in the work (through a new table Role).

Making this work required some small changes in various parts of the
codebase that were depending on the ad-hoc fields, amongst which the
most important probably are:
  - mangaki.management.commands.makefixtures, which was supposed to be a one-shot command and is removed in this patch
  - The `current_work_data` in the `Suggestion` model that was
    re-building a Work's template in-line inside the Django
    admin. This was replaced with a link "See on website" on the
    Suggestion's admin page.

This patch includes a migration that transforms the previous ad-hoc
fields into the new generic Role, but *does not remove them from the
database* yet; this will be done in a later patch. This ensures that a
rollback will be easy in case it is needed.


jilljenn:
Awesome work, thanks!

PR #178, opened 2016-05-12
Add ALS

pvigier:
I add als.py to train a model latent factors thanks to ALS. With 10 components, I obtain a RMSE of 0.21.

Surprisingly, I remarked that with mean shift, I obtain only a RMSE of 1.2, the same result obtained whith svd.py. I tried to disable mean shift in svd.py and I obtain a RMSE of 0.16 with 20 components.

I have no idea of why adding biases give poor results.

### DIFF
--- a/mangaki/mangaki/management/commands/compare.py
+++ b/mangaki/mangaki/management/commands/compare.py
@@ -24,7 +24,7 @@ class Experiment(object):
     results = {}
     algos = None
     def __init__(self, PIG_ID=None):
-        self.algos = [MangakiALS(), MangakiSVD(20), MangakiKNN()]
+        self.algos = [MangakiALS(20), MangakiSVD(20), MangakiKNN()]
         # self.results.setdefault('x_axis', []).append()
         self.make_dataset(PIG_ID)
         self.execute()
@@ -106,6 +106,6 @@ class Command(BaseCommand):
     help = 'Compare recommendation algorithms'
 
     def handle(self, *args, **options):
-        experiment = Experiment(1046)  # 1706
+        experiment = Experiment()  # 1706
         #experiment.display_ranking()
         # experiment.display_chart()
--- a/mangaki/mangaki/management/commands/compare.py
+++ b/mangaki/mangaki/management/commands/compare.py
@@ -5,6 +5,7 @@ from sklearn.cross_validation import train_test_split
 from mangaki.utils.svd import MangakiSVD
 from mangaki.utils.pca import MangakiPCA
 from mangaki.utils.knn import MangakiKNN
+from mangaki.utils.als import MangakiALS
 from mangaki.utils.values import rating_values
 from collections import Counter
 import numpy as np
@@ -23,7 +24,7 @@ class Experiment(object):
     results = {}
     algos = None
     def __init__(self, PIG_ID=None):
-        self.algos = [MangakiSVD(20)]
+        self.algos = [MangakiALS(), MangakiSVD(20), MangakiKNN()]
         # self.results.setdefault('x_axis', []).append()
         self.make_dataset(PIG_ID)
         self.execute()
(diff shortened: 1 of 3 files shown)


jilljenn:
On my data, which may be slightly different from yours, I get:

- ALS without mean shift: RMSE of 0.296
- SVD without mean shift: RMSE of 0.211
- SVD with mean shift: RMSE of 0.283
- KNN: RMSE of 0.472

But now comes something weird: even with NB_COMPONENTS = 1, I get the same RMSE for your code. There must be some bug lying somewhere. Let's investigate!

pvigier:
I investigate a little. My self.LAMBDA is very high. The values in the latent factors are very small (10^(-5)). Consequently, it predicts ratings of about 10^(-10).
```
[ALS]
Computing M: (1710 × 8347)
fill and center matrix [3639 ms]
Latent factors of user 0: [  4.06239871e-05   3.85350578e-05]
Latent factors of work 0: [  9.43476660e-06   8.98776766e-06]
Shapes (1710, 2) (2, 8347)
factor matrix [3816 ms]
Predictions: [  7.29629758e-10   8.28084715e-10  -2.95139991e-11 ...,   2.59517883e-10
   3.21587369e-10  -1.47224002e-10]
RMSE 0.205156948177
```
It is a bit weird to have such a low RMSE with these predictions.

jilljenn:
[ZERO]
RMSE 0.229162586265

:D

It must be because there is a lot of `willsee` and `wontsee` ratings. What about evaluating the models only over `like`, `dislike`, `favorite` and `neutral`, guys? :)

jilljenn:
Okay, got it. You should remove the value `1046` at the end of the `compare.py` file (it was a `PIG_ID`, for trying the experiment on a single user).

jilljenn:
I now get:

- ALS: 1.236
- SVD: 1.154
- KNN: 1.286
- ZERO: 1.544
- SVD without mean shift: 1.311

Fortunately, we should keep the mean shift!

mangaki/mangaki/management/commands/compare.py
```diff
@@ -106,5 +107,5 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         experiment = Experiment(1046)  # 1706
```
jilljenn: Here, you should remove 1046. Sorry :/

pvigier:
Nice, I now get:
```
[ALS]
RMSE 1.12438723474
[SVD]
RMSE 1.15403358119
[KNN] NB_NEIGHBORS = 15
RMSE 1.2924409726
```
The results are more consistent !

mangaki/mangaki/utils/als.py
```diff
@@ -0,0 +1,159 @@
...
+        self.M = self.U.dot(self.VT)
+
+        #self.save('backup.pickle')
+
+        self.chrono.save('factor matrix')
+
+    def predict(self, X):
+        return self.M[X[:, 0].astype(np.int64), X[:, 1].astype(np.int64)] + self.means[X[:, 0].astype(np.int64)]
+
+    def get_reco(self, username, sending=False):
```
jilljenn: You should remove this part for a cleaner code :) Please keep only what is useful for ALS.

pvigier:
I think, it's OK. Some suggestions ?

mangaki/mangaki/management/commands/compare.py
```diff
@@ -105,6 +106,6 @@ class Command(BaseCommand):
     help = 'Compare recommendation algorithms'
 
     def handle(self, *args, **options):
-        experiment = Experiment(1046)  # 1706
-        experiment.display_ranking()
+        experiment = Experiment()  # 1706
+        #experiment.display_ranking()
```
Elarnon: Is there a particular reason to remove this line?

Elarnon:
This looks good. Thanks for this alternative algorithm!

mangaki/mangaki/management/commands/compare.py
pvigier: It prints a lot (50 000) of predictions. It's not very convenient to examine the results of algorithms which are above, so I comment it.

jilljenn:
Congrats and welcome!

pvigier:
Thank you, very fun, I will follow your progress !

PR #194, opened 2016-06-26
Improve search bar aesthetics

Elarnon:


jilljenn:
Is it a normal behavior that search results are depending on the tab? (I assume so.) I don't get the same works in Most liked and Most popular.

Also I was a bit confused that I had to type Enter in order to trigger the search. Should there be some mention in the placeholder?

Awesome work anyway!

Elarnon:
The search results should be the same in all tabs, but the order may change : on the "most popular" tab they should be sorted by popularity etc. Maybe this is a wrong approach - but I believe that the current layout is ill-suited for searching anyways and should be re-thought to put a greater emphasis on searching (e.g. using tabs there doesn't really make much sense etc.).

Re: confusion about Enter, I will add a comment about this. In order not to have this information disappear once the user starts typing, I would rather have a help text stating "Don't forget to hit Enter or click on the search icon!".

jilljenn:
To be precise, if I type “manga”, I get some works in Most popular and Most controversial but no work in Most liked. It is because as I recall, in Most liked, some works were filtered. But you may have changed this part of the code.

Elarnon:
Ah, right. Yes, the filters are applied - so if the tab you are in filters out some works, they won't appear. Does resetting to the "Most popular" tab when searching (for now at least) sounds OK?

jilljenn:
Yes indeed. But the search query disappears (and again, I assume this is normal behavior).

PR #277, opened 2017-03-26
Enable anonymous user to rate works

Elarnon:
This is a first step towards the resolution of #59. It allows
anonymous users to rate works, and keeps the state of the rated works
across consecutive page loads in Django's session.

For now, the usage of this information is limited: although anonymous
user can rate works, they cannot get recommendations from those
ratings, transfer them to a new or existing account, etc. Support for
these will come in subsequent patches.

### DIFF
--- a/mangaki/mangaki/views.py
+++ b/mangaki/mangaki/views.py
@@ -101,7 +101,7 @@ def current_user_ratings(request, works=None):
     else:
         qs = user.rating_set.all()
         if works is not None:
-            qs = qs.filter(pk__in=works)
+            qs = qs.filter(work__in=works)
         return dict(qs.values_list('work_id', 'choice'))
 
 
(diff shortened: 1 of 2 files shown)


Elarnon:
Note: in general, this leaves all the parts that are currently not available to anonymous user (i.e. all views decorated with `@login_required`) for later patches.

mangaki/mangaki/views.py
```diff
@@ -60,6 +63,119 @@
...
+            string ('like', 'dislike', etc.)
+    """
+    user = request.user
+    if user.is_anonymous:
+        ratings = request.session.get(settings.ANONYMOUS_RATINGS_SESSION_KEY, {})
+        if works is not None:
+            filtered_ratings = {}
+            for work in works:
+                # Accept Work models as well as primary key integers
+                work = str(getattr(work, 'pk', work))
```
jilljenn: Un peu trop raccourci ?

Elarnon: Extracted into a `pk_from_object_or_pk` method in the following patch in the series (#278).

PR #278, opened 2017-03-26
Enable anonymous users to get recommendations

Elarnon:
This is a second step towards the resolution of #59. It allows
anonymous users to get recommendations (using the KNN algorithm only,
since we do not have trained models for them).

This is implemented by making the various algorithms take directly a
list or (user_id, work_id, choice) triplets instead of building those
from a queryset, as well as some glue to add the relevant triplets
from anonymous users at the interface between the database and the
algorithms themselves.

### DIFF
--- a/mangaki/mangaki/views.py
+++ b/mangaki/mangaki/views.py
@@ -18,12 +18,14 @@ from django.utils.functional import cached_property
 from django.db.models import Case, When, Value, Sum, IntegerField
 from django.views.generic.detail import SingleObjectMixin
 from django.db import connection, DatabaseError
-from django.conf import settings
 
 from mangaki.models import Work, Rating, ColdStartRating, Page, Profile, Artist, Suggestion, Recommendation, Pairing, Top, Ranking, Staff, Category, FAQTheme, Trope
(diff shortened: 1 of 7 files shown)


mangaki/mangaki/views.py
```diff
@@ -18,12 +18,14 @@
 from django.db.models import Case, When, Value, Sum, IntegerField
 from django.views.generic.detail import SingleObjectMixin
 from django.db import connection, DatabaseError
-from django.conf import settings
 
 from mangaki.models import Work, Rating, ColdStartRating, Page, Profile, Artist, Suggestion, Recommendation, Pairing, Top, Ranking, Staff, Category, FAQTheme, Trope
 from mangaki.mixins import AjaxableResponseMixin, JSONResponseMixin
 from mangaki.forms import SuggestionForm
 from mangaki.utils.mal import import_mal
+from mangaki.utils.ratings import (
```
jilljenn: Why brackets? For the 80 chars?

Elarnon: I'd have said "for readability", but essentially, yeah.

PR #315, opened 2017-05-11
Upgrade to Django 1.11

Elarnon:
Looks like we are not using any (broken things)[https://docs.djangoproject.com/en/1.11/releases/1.11/#backwards-incompatible-changes-in-1-11] in Django 1.11, although I only did a simple test with a local `runserver`.

### DIFF
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -1,4 +1,4 @@
-Django==1.10.*
+Django==1.11.*
 django-allauth==0.28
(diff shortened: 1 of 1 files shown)


jilljenn:
Did you test browsing the website or just running the server? :smile:

Elarnon:
I tried browsing the website, although only simple interactions (rating a few works, etc.).

PR #323, opened 2017-05-21
[Hotfix] Handle recommended ghost works

jilljenn:
To reproduce bug:

- Get SVD recommendations
- Delete one of those works
- Crash.

### DIFF
--- a/mangaki/mangaki/views.py
+++ b/mangaki/mangaki/views.py
@@ -655,9 +655,8 @@ def get_reco_algo_list(request, algo, category):
     data = get_reco_algo(request, algo, category)
     works = data['works']
     for work_id in data['work_ids']:
-        work = works.get(work_id)
-        if work is not None:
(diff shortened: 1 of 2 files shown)


mangaki/mangaki/views.py
```diff
@@ -655,8 +655,9 @@ def get_reco_algo_list(request, algo, category):
     data = get_reco_algo(request, algo, category)
     works = data['works']
     for work_id in data['work_ids']:
-        work = works[work_id]
-        reco_list.append({'id': work.id, 'title': work.title, 'poster': work.ext_poster, 'synopsis': work.synopsis,
+        work = works.get(work_id)
```
Elarnon: Not the right place to make this check IMO. The filtering should happen at the end of `get_reco_algo`, which shouldn't return recommendations for non-existent works, ever.

PR #356, opened 2017-06-24
Add i18n in English and Japanese

jilljenn:
Only deployed for about.html but I thought it was nice to start with good practices.

FR
EN
JP


mangaki/mangaki/templates/base.html
```diff
@@ -59,35 +61,45 @@
                 <div class="collapse navbar-collapse" id="menu-collapse">
                     <ul class="nav navbar-nav">
                         <li class="dropdown">
-                            <a href="{% url 'about' %}">
+                            <a href="{% url 'about' 'fr' %}">
                                 C'est quoi ?
                             </a>
                         </li>
+                        <li class="dropdown">
```
RaitoBezarius: In fact, I think we are doing it wrong.

We should rely on `Accept-Language` header to show the good `What is it` rather than showing them *all*.

jilljenn: And how to change language?

RaitoBezarius: We should add a button in the footer to help people, like dissem.in does.

jilljenn:
Good right?
capture d ecran 2017-07-01 a 02 15 14

RaitoBezarius:
Better, indeed @jilljenn 
But, I'd just add a language icon / flag so that it can be clear crystal for users.
You have conflicts to resolve, @jilljenn

PR #414, opened 2017-07-13
AniDB's API wrapper fetches related animes

mrsalixor:
This PR adresses the final list item of issue #382 by retrieving related animes.
This requires a new model to store relations between works : `RelatedWork`.

AniDB has different types of relations : *Prequel, Sequel, Summary, Side Story, Parent Story, Alternative Setting, Same Setting or Other*.

My `RelatedWork` model follows this structure :
- `pk`, the primary key of the entry in the database
- `parent_work`, the ID of the *parent work* ie. the one which relations are being looked at
- `child_work`, the ID of the *child work* ie. the one which the `type` attribute refers to
- `type`, that takes a value as seen before, formatted appropriately

For instance, if I have three works in my database, *Season 1* (`work_id=1`), *Season 2* (`work_id=2`) and an *OVA* (`work_id=3`), then the `RelatedWork` model would look like (after appropriate operations) :
```
 pk | parent_work_id | child_work_id |   type
-----------------------------------------------
  1 |        1       |       2       | 'sequel'
  2 |        1       |       3       | 'other'
```


mangaki/mangaki/models.py
```diff
@@ -269,8 +269,11 @@ def is_nsfw_based_on_tags(self, anidb_tags):
         sum_weight_nsfw_high = sum(infos["weight"] for t, infos in anidb_tags.items()
                                    if t in potentially_nsfw_tags and infos["weight"] > 400)
 
-        percent_low_nsfw = (sum_weight_nsfw_low/sum_weight_all_low) * 100
-        percent_high_nsfw = (sum_weight_nsfw_high/sum_weight_all_high) * 100
+        if sum_weight_all_low > 0 and sum_weight_all_high > 0:
```
mrsalixor: I'll move stuff unrelated to this PR to another PR 😉

mangaki/mangaki/utils/anidb.py
```diff
@@ -17,7 +17,9 @@ def to_python_datetime(mal_date):
     >>> to_python_datetime('2015-07-14')
     datetime.datetime(2015, 7, 14, 0, 0)
     """
-    return datetime(*list(map(int, mal_date.split("-"))))
+    date = list(map(int, mal_date.split("-")))
+    date += [1 for _ in range(3-len(date))]
```
mrsalixor: Yeah, the issue is that AniDB only gives us the year an anime is planned to release. When it's the case, an error is raised since the format is not that which is expected.

PR #421, opened 2017-07-17
Retrieve a user's anime or manga list from AniList

mrsalixor:
The AniList's API wrapper can now fetch a user's manga or anime list.
I also added a `anilist_id` field to the Work model but I'm not sure if it really has its place in this PR.

(*Refers to issue #343*)


mrsalixor:
I removed the new `anilist_id` field from this PR, we could add it back in another PR if needed !

PR #460, opened 2017-08-09
mangaki/fix interface

mrsalixor:
This PR adds a new mangaki/fix interface for `Suggestion` objects.

A new `Evidence` model, to indicate whether a `User` agrees with a `Suggestion`, and whether they think it requires the help of an administrator.
*An `Evidence` object is created when an user approves or disapproves a `Suggestion`.*

A new `origin` field on `WorkCluster`, to link a cluster and a `Suggestion` (*not required*)

-------------
Right now, there are two new pages :
- **`mangaki.fr/fix` :** list `Suggestion` objects and display some information (*for now, ordered by most recent to oldest*)
- **`mangaki.fr/fix/suggestion/<id>` :** display a `Suggestion` object (and the related `WorkCluster` if exists) with information and two toggles – `agree` and `needshelp`


mangaki/mangaki/views.py
```diff
@@ -807,6 +809,78 @@ def legal_mentions(request):
     return render(request, 'mangaki/legal.html')
 
 
+def fix_index(request):
+    suggestion_list = Suggestion.objects.select_related('work', 'user').all().prefetch_related(
```
RaitoBezarius: I'm not sure but I think that you could put `prefetch_related` and `order_by` before the `all` call.

mangaki/mangaki/views.py
```diff
@@ -807,6 +809,90 @@ def legal_mentions(request):
...
+        'evidence': evidence,
+        'next_id': next_suggestion.id if next_suggestion else None,
+        'previous_id': previous_suggestion.id if previous_suggestion else None
+    }
+
+    return render(request, 'fix/fix_suggestion.html', context)
+
+
+def update_evidence(request):
+    agrees = request.POST.get('agrees')
```
RaitoBezarius: That'd be good if we could use instead Django Forms for simplifying the code.

mangaki/mangaki/views.py
```diff
@@ -807,6 +809,90 @@ def legal_mentions(request):
...
+    }
+
+    if not suggestion_id:
+        return redirect('fix-index')
+
+    suggestion = get_object_or_404(Suggestion.objects.select_related('work', 'user'), id=suggestion_id)
+
+    evidence = None
+    if request.user.is_authenticated:
+        evidence = Evidence.objects.filter(user=request.user, suggestion=suggestion).first()
```
RaitoBezarius: Do we have any reason to not use `get` and enforce the uniqueness of Evidence per `(user, suggestion)` ? (for the non-existence, we can catch the DoesNotExist exception and set `evidence` to `None` or whatever).

mangaki/mangaki/views.py
```diff
@@ -807,6 +809,90 @@ def legal_mentions(request):
...
+
+    cluster = WorkCluster.objects.filter(origin=suggestion_id).first()
```
RaitoBezarius: Same here, can we have more than one work cluster with `origin=suggestion_id` ?

mangaki/mangaki/views.py
```diff
@@ -807,6 +809,90 @@ def legal_mentions(request):
...
+    if not next_suggestion and not suggestion.is_checked:
+        next_suggestion = Suggestion.objects.filter(is_checked=True).order_by('-date').first()
+
+    context = {
+        'suggestion': suggestion,
+        'related_cluster': cluster,
+        'cluster_works': cluster_works,
+        'cluster_colors': cluster_colors[cluster.status] if cluster else None,
+        'evidence': evidence,
+        'next_id': next_suggestion.id if next_suggestion else None,
```
RaitoBezarius: Let's fetch ID from the start using `values_list` rather than fetching the whole `suggestion`

mangaki/mangaki/views.py
mrsalixor: I updated and used a Django Form for updating or creating an `Evidence` object ;)
However but I'm not sure where to move the `update_evidence` method (if it has to be moved).

I also updated the form so that it can be used if the user doesn't have JavaScript enabled.

mangaki/mangaki/templates/fix/fix_index.html
```diff
@@ -0,0 +1,52 @@
...
+<h1>Liste des suggestions</h1><br/>
+
+{% for suggestion in suggestions %}
+<div class="row">
+<div class="panel panel-default">
+    <h3 class="panel-heading panel-title">
+        <a href="{% url "fix-suggestion" suggestion.id %}" class="text-primary">Suggestion #{{ suggestion.id }}</a>
+        <span class="small font-weight-bold">
+            &nbsp;&nbsp;<span class="glyphicon glyphicon-{% if suggestion.is_checked %}lock{% else %}flag{% endif %}" aria-hidden="true"></span>
+            Ce problème est {% if suggestion.is_checked %}fermé{% else %}en cours{% endif %}
```
RaitoBezarius: I think the terminology behind `is_checked` is more like:
- an admin has checked a suggestion and valid its contents and triggered actions (or not)
So rather than "fermé", "résolu" seems better?

mangaki/mangaki/templates/fix/fix_suggestion.html
```diff
@@ -0,0 +1,143 @@
...
+
+<div class="col-md-9">
+    <br/>
+    <div class="row">
+    <div class="col-md-push-1 col-sm-push-1 col-lg-10 col-sm-10">
+        <div class="input-group">
+            <span class="input-group-addon" id="basic-addon2">
+                <span class="glyphicon glyphicon-link"></span> Lien de partage
+            </span>
+            <input type="text" class="form-control" value="{{ request.build_absolute_uri }}">
```
RaitoBezarius: Maybe, let's add a `{# TODO: use JS to select the contents and put it in the clipboard when clicking on the field #}` ?

mangaki/mangaki/templates/fix/fix_suggestion.html
```diff
@@ -0,0 +1,143 @@
...
+                {% endfor %}
+            </div>
+            <hr/>
+            <p>
+                Ce cluster est <strong><span class="{{ color }}">{{ cluster.get_status_display.lower }}</span></strong>
+                {% if cluster.checker %} par <a href="{% url "profile" cluster.checker %}">{{ cluster.checker }}</a>{% endif %}.
+            </p>
+            {% if cluster.resulting_work %}
+            <p>
+                Ce cluster a résulté en l'œuvre
```
RaitoBezarius: A bit weird, we could say: "Ce cluster a été fusionné dans l'œuvre:" (but I'm not convinced also by my terminology… hahaha, if you have any ideas, don't hesitate!)

mangaki/mangaki/templates/fix/fix_suggestion.html
mrsalixor: I found a handy [CodePen](https://codepen.io/nathanlong/pen/ZpAmjv) so I used it to make it possible to copy the link ;)

RaitoBezarius:
So far I have been testing, I have two questions:

- Shouldn't we register `Evidence` in the `admin.py` file for administrators?
- It seems like the `Evidence` is not saved when clicking on the "approve" button.

Niceties:

- Maybe we should add on each suggestion the amount of "dis/approved suggestions" on the detail page and on the list page?
- In the description of a suggestion, we could parse the links to make them true links (but let's be wary of HTML! Escape everything!) — good job, no easy XSS: https://imgur.com/r4YTsW6
- I'm still looking at stuff, maybe we should add a "Suggestions" tab in the general navigation menu?

RaitoBezarius:
Another question:

- In fact, it's the switch "approve / disapprove" could be a bad idea for the first time.
We should explicitly show an :+1: or :-1: and then show a switch.
Because, it'd mislead the user into thinking that he disapproves by default (when in fact, he has not expressed himself) 

UI/UX nitpicks (maybe for another PR!):
Also, the "need help" should be a one time link, like "J'ai besoin d'un administrateur sur cette suggestion" and will become "Un administrateur a été demandé!" (I don't think you want to revert your "need help", but well, you could by showing a small "annuler" but we should make it really difficult to click unintentionally).

mrsalixor:
I updated the UI for approving/disapproving a `Suggestion` !

Rather than using a toggle, I'm now using a 👍/👎 approach :
- I deleted the `EvidenceForm` since it's no longer used.
- I removed AJAX calls since it doesn't feel needed anymore (but I can try to add some back if you feel it would be better with some AJAX !)

I also made it possible to delete an `Evidence`, and *impossible* to cancel a call for an admin. *(Well, it's still possible if the user deletes their `Evidence`)*

RaitoBezarius:
Now, the PR looks very hot, I must say.

Some screens for others:

A suggestion detail page featuring Chuunibyou demo koi ga Shitai!
A list of suggestions

mangaki/mangaki/templates/fix/fix_index.html
```diff
@@ -0,0 +1,69 @@
...
+<h1>Liste des suggestions</h1><br/>
+
+{% for suggestion in suggestions %}
+<div class="row">
+<div class="panel panel-default">
+    <h3 class="panel-heading panel-title">
+        <a href="{% url "fix-suggestion" suggestion.id %}" class="text-primary">Suggestion #{{ suggestion.id }}</a>
+        <span class="small font-weight-bold">
+            &nbsp;&nbsp;<span class="glyphicon glyphicon-{% if suggestion.is_checked %}lock{% else %}flag{% endif %}" aria-hidden="true"></span>
+            Ce problème est {% if suggestion.is_checked %}résolu{% else %}en cours{% endif %}
```
RaitoBezarius: It can be rewritten as: `{{ suggestion.is_checked |yesno:'résolu,en cours' }}` (according to [this](https://docs.djangoproject.com/fr/1.11/ref/templates/builtins/#yesno)).

mangaki/mangaki/views.py
```diff
@@ -807,6 +809,120 @@ def legal_mentions(request):
...
+        'evidence': evidence,
+        'next_id': next_suggestions_ids[0] if next_suggestions_ids else None,
+        'previous_id': previous_suggestions_ids[0] if previous_suggestions_ids else None
+    }
+
+    return render(request, 'fix/fix_suggestion.html', context)
+
+
+def update_evidence(request):
+    result = (request.POST if request.method == 'POST'
```
RaitoBezarius: In which circumstances, we update an evidence without a `POST`?

mangaki/mangaki/views.py
```diff
@@ -807,6 +809,120 @@ def legal_mentions(request):
...
+        'clusters': zip(clusters, colors) if clusters and colors else None,
+        'evidence': evidence,
+        'next_id': next_suggestions_ids[0] if next_suggestions_ids else None,
+        'previous_id': previous_suggestions_ids[0] if previous_suggestions_ids else None
+    }
+
+    return render(request, 'fix/fix_suggestion.html', context)
+
+
+def update_evidence(request):
```
RaitoBezarius: I think we should add [`@login_required`](https://docs.djangoproject.com/fr/1.11/topics/auth/default/#django.contrib.auth.decorators.login_required) here, in fact, we should not be able to call this method if we are not logged in.

PR #569, opened 2018-06-01
Fix KNN reco for hipsters

### LINKED ISSUE #562: Si des gens ont liké des œuvres que personne d'autre n'a likées, alors ils ne peuvent pas obtenir de recommandations
À cause de la ligne :

    framed_rated_works['work_id'] = dataset.encode_works(framed_rated_works['work_id'])

Du coup, que faire.

La solution : faire un `df.query('work_id in @available_works')` au préalable, où `available_works` est la liste des clés de `dataset.encode_work`.

jilljenn:
Hipsters == people who like anime that were not in the algorithm snapshot (pickle).

Fixes #562.


jilljenn:
L'ascension du coverage continue ! 🚀

PR #594, opened 2018-08-12
Make English default language, remove a lot of text to ease translation

jilljenn:


jilljenn:
Almost finished! Now the hard parts:

- Some sentences (few) are in the JavaScript code
- Every time there are HTML tags or, worse, template tags, of course Google Translate struggles.

PR #620, opened 2018-09-04
Update simplified Chinese translation

xuedong:


xuedong:
There could be errors, but I actually need a place for further discussion, on some context problems for e.g., that's why I created this pull request in the first place.

jilljenn:
Sorry, I was busy with the AAAI deadline 😄 Wonderful!
Actually, I wanted to suggest to you to do it gradually. By minibatches 😄 Don't you want us to merge the current one?

xuedong:
I corrected some errors, and updated some new translations, and I think now this django.po for simplified Chinese is more or less done. However I also noticed some inconsistent phrases among different languages (not only in Chinese). And I will be focusing on AISTATS as well xD, so maybe we can discuss that later, and we can merge the current commit for now?

PR #626, opened 2018-09-17
Add online KNN, SVD and ALS

jilljenn:
🚀

I also fixed the `pycodestyle` of this file.


jilljenn:
Seems like the time for recommendations on my laptop (which is faster than the production):
- before: 443+1+207+5 = 656ms
- now knn: 455+3+25+5 = 488ms
- now svd: 122+1+6+4 = 133ms
- now als: 119+1+4+5 = 129ms
For someone with 11 ratings.

PR #652, opened 2019-02-22
Version link to the release page.

### LINKED ISSUE #602: Version 0.6 should link to the releases page

FardaleM:
fix #602

### DIFF
--- a/mangaki/mangaki/templates/base.html
+++ b/mangaki/mangaki/templates/base.html
@@ -193,7 +193,7 @@
             </ul>
             <ul class="navbar-nav">
                 <li class="nav-item">
-                    <a class="nav-link" href="https://github.com/mangaki/mangaki/commits/{% mangaki_revision %}">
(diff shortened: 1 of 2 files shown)


RaitoBezarius:
This breaks the "development version" thing and will give less information than before, what must be done is to auto-detect when a certain version of Mangaki is in fact a release and when it's not, I think that'd be the most optimal way.

jilljenn:
@RaitoBezarius I feel that anyway, currently the link to commit is 404, so it's better to merge the PR. Okay?

RaitoBezarius:
I prefer for it to be 404, it's okay for now because it's a developer
feature anyway.

jilljenn:
:/
I don't like 404 on the production website https://mangaki.fr, it's not good for SEO as well; at least let's remove the link?

RaitoBezarius:
I'm okay on commenting the link, but no removal. It's a broken feature, not
an obsolete one I believe.

jilljenn:
OK sorry @FardaleM !

PR #657, opened 2019-07-30
Add django-massmailer integration

seirl:
The style of some parts is broken but that's because your CSS is weird.

### DIFF
--- a/mangaki/mangaki/settings.py
+++ b/mangaki/mangaki/settings.py
@@ -69,7 +69,11 @@ INSTALLED_APPS = (
     'bootstrap4',
     'django_js_reverse',
     'rest_framework',
-    'django_celery_beat'
+    'django_celery_beat',
+    'massmailer',
+    'crispy_forms',
+    'django_bootstrap_breadcrumbs',
+    'reversion',
 )
 
 if config.has_section('sentry'):
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -4,6 +4,7 @@ django-allauth==0.38.0
 django-bootstrap4
 django-celery-beat==1.1.*
 django-js-reverse==0.8.2
+django-massmailer
 django-sendfile==0.3.*
 djangorestframework==3.8.2
 psycopg2-binary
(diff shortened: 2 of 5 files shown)


requirements/production.txt
```diff
@@ -4,6 +4,7 @@ django-allauth==0.38.0
 django-bootstrap4
 django-celery-beat==1.1.*
 django-js-reverse==0.8.2
+django-massmailer
```
RaitoBezarius: Can you pin a version?

mangaki/mangaki/settings.py
```diff
@@ -69,7 +69,11 @@
     'bootstrap4',
     'django_js_reverse',
     'rest_framework',
-    'django_celery_beat'
+    'django_celery_beat',
+    'massmailer',
+    'crispy_forms',
+    'django_bootstrap_breadcrumbs',
+    'reversion',
```
RaitoBezarius: What is reversion?

requirements/production.txt
seirl: Considering I'm still unbreaking stuff on it, I don't really advise it. You're the first early adopters!

mangaki/mangaki/settings.py
seirl: A dependency of massmailer that is used to store past versions of the templates and queries: https://github.com/etianen/django-reversion

seirl:
> Sounds good, what about the "weird CSS" stuff? Can you elaborate or be more precise? Can we fix it on our own?

It's basically because your CSS is too broad, so it changes stuff that don't have class attributes, which breaks some parts of the design.

seirl:
(Please note the draft status of this, I'm still making some changes and looking for early feedback, don't merge this)

RaitoBezarius:
Yes, I understood, I was taking an early look to it as I was curious. Thank
you for this !

RaitoBezarius:
Closed in favor of #765 which will use listmonk for now.
Deep-integration with Mangaki is yet to be done, but unnecessary for now.

PR #670, opened 2019-10-27
Add missing h1 tags

jilljenn:
Let's try to fix our broken SEO.

Also some pages have two h1 tags, for example in the preferences of the profile.


jilljenn:
Plz @RaitoBezarius!

PR #690, opened 2020-06-05
Fix Vagrant setup

Elarnon:
This mostly changes the Ansible configuration to be compatible with Vagrant:

 - Use Debian Buster
 - Enable pipelining in ansible.cfg
   (see https://docs.ansible.com/ansible/latest/user_guide/become.html#risks-of-becoming-an-unprivileged-user )
 - Use python3 for starting ansible on the host (default interpreter_python
   in ansible.cfg)
 - Some cleanup based on Ansible warnings (mostly stop using loops
   for `apt`)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,7 +14,7 @@ Vagrant.configure(2) do |config|
     # For a complete reference, please see the online documentation at
     # https://docs.vagrantup.com.
     #
-    config.vm.box = "debian/contrib-jessie64"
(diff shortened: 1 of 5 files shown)


Elarnon:
> Some cleanup based on Ansible warnings (mostly stop using loops for apt)

J'ai eu des warnings disant "blabla il faut pas faire des with_items avec apt il faut donner une liste à l'argument name" alors moi j'ai suivi les instructions.

